PR #12, opened 2023-06-19
Cleaning tools and "setup and run"

Jgmedina95:


SamCox822 (commented):
In mdagent/tools/search_tools.py
can you change the prompt summary to use the BaseLanguageModel from langchain (it's also used in tools.py). We need to keep llm model types the same throughout

mdagent/tools/search_tools.py
SamCox822, line 1: Can you change the prompt summary to use the BaseLanguageModel from langchain (it's also used in tools.py). We need to keep llm model types the same throughout

notebooks/set_and_run_example.ipynb
SamCox822, line 1: Can you make this a full example that makes the params file and then runs?

environment.yaml
SamCox822, line 1: Since you have to add openmm and pdbfixer via conda, can you add those instructions on the readme under Installation?

SamCox822 (changes requested):

mdagent/tools/search_tools.py
Jgmedina95: done

notebooks/set_and_run_example.ipynb
Jgmedina95: done

environment.yaml
Jgmedina95: done

SamCox822 (commented):
I did a little rearranging and added a notebook that reflects a full user story.

SamCox822 (approved):

PR #13, opened 2023-06-20
implement name2pdb tool

qcampbel:
* added name2pdb tool
* added test
* example notebook 

Note: don't approve yet - working on the tool and its description so the agent won't use literature search unintentionally


Jgmedina95:
Issue with notebook. For some reason the agent is not importing the tool

qcampbel:
By rerunning the updated notebook on my end, the agent imported the tool successfully. It was able to create PDB file. However, the agent would run it multiple times before giving PDB file as the final answer.. It can be approved as is, since it completed its task, or

Jgmedina95 (approved):
After changing the return statement, is worked like a charm

PR #14, opened 2023-06-22
automate loading API keys

qcampbel:
Make it easier and cleaner for developers to load API keys. Instead of adding the keys directly in each of notebooks, one just need to create .env file with API keys and run notebooks as is.
* sample_env is provided as a template for .env file
* .env isn't mandatory to run the agent (API keys can be added manually by users as before), but it'll keep keys secure, as .env files won't get pushed to the repo


Jgmedina95:
There are not new tests checking this new configuration. it doesn't need it?

Is it up to us to create the .env file in our setup?, I've seen other implementations where a .env.example file is used to make it easier for new developers.

SamCox822:
> There are not new tests checking this new configuration. it doesn't need it?
> 
> Is it up to us to create the .env file in our setup?, I've seen other implementations where a .env.example file is used to make it easier for new developers.

@Jgmedina95 no, we shouldn't need tests for this other than the notebooks. This isn't a new function, it just adds an optional simpler/cleaner method. This PR will create an example file that users can edit/update appropriately for their use. We can think about a potential better way to include this, but for now, I think it's fine.

SamCox822 (dismissed):

PR #22, opened 2023-06-28
Rmsd

qcampbel:
* implement rmsd functions  using MDanalysis
* also improved name2pdb so it can produce pdb files by default, since MDAnalysis doesn't work with .cif files. Name2pdb now can produce either cif or pdb files.


qcampbel:
decided to convert this PR to draft until I add RMSD plotting

Jgmedina95:
The notebook works!. My last worry would be the description of the ppi tool:  
```
"  Calculates minimum heavy-atom distance between peptide (assumed to be
    smallest chain) and protein. Returns average distance between these two."
```
As it actually calculates and returns the average between all residues of the protein. Maybe just add protein _residues_ in the description.

Jgmedina95 (approved):

PR #43, opened 2023-11-07
Implement VectorDB and simplify adding new tools

qcampbel:
Changes made:

- implemented Chroma vector database storing skills for SkillRetrieval
- implement Chroma vector database storing all tools for tool retrieval when initializing MDAgent
- Simplified creating Langchain tools and executing them by making it more 'function-centered'
- pass all_tools and current_tools to Iterator code (still need to have other agents to include those new inputs)

I also updated subagent tools and agent type by switching to OpenAI Functions type


Jgmedina95 (commented):
Some comments:
"python_repl" tool to be changed from langchain_experimental too. 
agent.py:

1.  Why there are two initialization of the agent_executor?
2. Also, the prompt initialization for openai agent is different than with MRKL. It needs something like this per the documentation:

```
            prompt = ChatPromptTemplate.from_messages(
                [
                    ("system", "\n\n".join([PREFIX, FORMAT_INSTRUCTIONS_FUNC])),
                    ("user", QUESTION_PROMPT_FUNC),
                    MessagesPlaceholder(variable_name="agent_scratchpad"),
                ]
            )
```
Otherwise, it just gets ignored. 

in general, inside the subagents. A little more of comments explaining each function would be helpful :)

qcampbel:
1. I cleaned up agent initialization so it's initialized once under 'run' function. Because only tools relevant to the prompt are retrieved, it has to be initialized after prompt & a list of tools are obtained.

Jgmedina95 (approved):

PR #52, opened 2023-12-18
2 agent options

SamCox822:


SamCox822:
the goal here @qcampbel @Jgmedina95 is to merge this into updatingsimul2 before approving & merging that one

qcampbel (approved):
Looks good. if @Jgmedina95 agrees, let's merge. 

I notice mypy.ini was deleted though. was it intended?

Jgmedina95 (approved):

SamCox822:
@qcampbel 

> Looks good. if @Jgmedina95 agrees, let's merge.
> 
> I notice mypy.ini was deleted though. was it intended?

yes, we added the mypy flag `--ignore-missing-imports` a while back to the pre-commit config file, so we don't need the mypy init file anymore

PR #62, opened 2024-01-03
update CreateNewTool to execute as well

qcampbel:
- partly combine CreateNewTool and ExecuteSkill, since it makes more sense to do it in one step instead of two.
- renamed ExecuteSkill -->  RetryExecuteSkill. Just in case executing fails (likely due to missing arguments). May remove this later if deemed unnecessary.


mdagent/tools/maketools.py
```diff
@@ -186,15 +185,21 @@ class CreateNewToolInputSchema(BaseModel):
         description="""List of all tools you have access to. Such as
         this tool, 'ExecuteSkill', 'SkillRetrieval', and maybe `Name2PDBTool`, etc."""
     )
+    execute: Optional[bool] = Field(
+        description="Whether to execute the new tool or not. default: True"
+    )
```
Jgmedina95, line 190: you can include the default value like: 
```{python}
Field(True, description="Whether to execute the new tool or not. default: True")
```
That way, you don't need to specify the value later on when calling _run

Jgmedina95 (commented):
See comment in maketools file

mdagent/tools/maketools.py
qcampbel: updated in latest commits

Jgmedina95 (commented):
LGTM

Jgmedina95 (approved):

PR #63, opened 2024-01-03
bring back tool retrieval

qcampbel:
- Added back tool retrieval for the main dude
- Also improved how tools are retrieved to make it more robust and useful

@SamCox822 and @Jgmedina95 - look at agent.py and tell me what you think. Old version of tool retrieval looks awkward with agent initialized during 'run' function, so I made a separate function for agent initialization to make it look cleaner and readable.


Jgmedina95:
Im confused here, why do we want to initialize the agent with the run call? doesn't this will make it impossible to have memory as the agent will reinitialize after every query?

SamCox822 (approved):

PR #66, opened 2024-01-04
Improving path registry

Jgmedina95:
1. Give value to the path_registry. Files names are now files id. Which are the ones the agent will use to use files inside tools (already implemented in simulations and cleaning tools)
2. Improve ListRegistryPath as names are hardly self-explanatory, a description of each file is also included.

### DIFF
--- a/tests/test_fxns.py
+++ b/tests/test_fxns.py
@@ -1,3 +1,4 @@
+import json
 import os
 import warnings
 from unittest.mock import MagicMock, mock_open, patch
@@ -11,7 +12,7 @@ from mdagent.tools.base_tools import (
     get_pdb,
 )
 from mdagent.tools.base_tools.analysis_tools.plot_tools import plot_data, process_csv
-from mdagent.utils import PathRegistry
+from mdagent.utils import FileType, PathRegistry
 
 warnings.filterwarnings("ignore", category=DeprecationWarning, module="pkg_resources")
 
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
qcampbel, line 1: I think this file is accidentally being pushed to this repo

mdagent/tools/maketools.py
```diff
@@ -80,7 +79,7 @@ def make_all_tools(
         CheckDirectoryFiles(),
         #    InstructionSummary(path_registry=path_instance),
         ListRegistryPaths(path_registry=path_instance),
-        MapPath2Name(path_registry=path_instance),
+        #    MapPath2Name(path_registry=path_instance),
         Name2PDBTool(path_registry=path_instance),
```
qcampbel, line 83: you sure we want to remove that tool? suppose there are some files provided by users - just a thought

.vscode/settings.json
Jgmedina95: youre right! it got in accidentally when I pulled the last PR

mdagent/tools/maketools.py
Jgmedina95: good point, is just that the agent sometimes adds the same file more than once with the tool taking it made it work better for my examples.

SamCox822 (commented):
can you add unit tests for the new features you added?

mdagent/tools/base_tools/preprocess_tools/clean_tools.py
```diff
@@ -343,26 +350,41 @@ def _run(self, **input_args) -> str:
...
-                file_name = f"tidy_{name}v{version}.{end}"
-
+            file_name = self.path_registry.write_file_name(
+                type=FileType.PROTEIN,
+                protein_name=name,
+                description="Clean",
+                file_format=end,
+            )
+            file_id = self.path_registry.get_fileid(file_name, FileType.PROTEIN)
+            #            if output_path:
```
SamCox822, line 360: is there a reason we are leaving these comments in?

mdagent/tools/maketools.py
```diff
@@ -80,7 +79,7 @@ def make_all_tools(
         CheckDirectoryFiles(),
         #    InstructionSummary(path_registry=path_instance),
         ListRegistryPaths(path_registry=path_instance),
-        MapPath2Name(path_registry=path_instance),
+        #    MapPath2Name(path_registry=path_instance),
```
SamCox822, line 82: and this comment

SamCox822 (approved):

PR #74, opened 2024-01-29
added streamlit app with ability to upload files

SamCox822:
just for fun 

This PR, specifically:
- Instructions added to readme
- streamlit package added to setup
- some logging messages duplicated (previously just printed to terminal - now also to streamlit app)
- streamlit app added -> when files are uploaded, they get added to registry

Things we can do in the future: 
- more add st.markdown lines to print directly to the app as needed
- add ability to handle different file types
- implement chat model & use streamlit app to chat

### DIFF
--- a/mdagent/tools/base_tools/simulation_tools/setup_and_run.py
+++ b/mdagent/tools/base_tools/simulation_tools/setup_and_run.py
@@ -317,8 +317,8 @@ class SimulationFunctions:
             ]
             Forcefield = Forcefield_files[0]
             Water_model = Forcefield_files[1]
-        print("Setting up forcields :", Forcefield, Water_model)
-        st.markdown("Setting up forcields", unsafe_allow_html=True)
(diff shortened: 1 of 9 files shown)


mdagent/tools/base_tools/simulation_tools/setup_and_run.py
brittyscience, line 1: Please check line 321, I think I noticed a typo.

brittyscience: @SamCox822

brittyscience (changes requested):
I am unclear if this is a "PR" or a simple review of a commit. Did you mean .CSV or .CIF file? The streamlit was tested and run, it appears that .CIF is the standard format from PDB.

brittyscience (approved):
I have witnesses. PR approved :)

PR #76, opened 2024-01-31
trying to use rdkit for small mol pdb

SamCox822:
- a new tool that takes in a mol name/smiles and creates a pdb file for it using rdkit
- Packmol now explicitly takes in a list of small molecules, which are downloaded if not already 
- unit tests for the download pdb and the download pdb part of Packmol tool


Jgmedina95:
Just went through it quickly. Looks very good, and I think is an improvement. It also reduces the dependency on path_registry to work all the time. Ill test it later today but couple questions to think about: 
1. Should we remove solvents from the path_registry then? i differentiate in different directories cause pdb was intended for proteins and solvents for single small molecules that are used as solvents. 
2. I don't know if this is overthinking, but I'm concerned the agent will think a small protein (ALA I think?) for example is a small molecule too? maybe include this in some more detail in the tool description. 
3. I noted your using document strings, """ """, in descriptions and return statements. I would like to change those in ("") as it formates prettier inside the agent (as explained before in the slack channel).
4. Again, love the idea hjaha I'm going to test it later today after couple meetings :)

SamCox822:
> Just went through it quickly. Looks very good, and I think is an improvement. It also reduces the dependency on path_registry to work all the time. Ill test it later today but couple questions to think about:
> 
> 1. Should we remove solvents from the path_registry then? i differentiate in different directories cause pdb was intended for proteins and solvents for single small molecules that are used as solvents.

I don't think we need the folders? what do you think? 

> 2. I don't know if this is overthinking, but I'm concerned the agent will think a small protein (ALA I think?) for example is a small molecule too? maybe include this in some more detail in the tool description.

I don't think this is particularly concerning. I doubt this will happen, and if it does, it will either realize its mistake via error messages or end up using the small molecule route to get a pdb file...which should still be a valid pdb file.

> 3. I noted your using document strings, """ """, in descriptions and return statements. I would like to change those in ("") as it formates prettier inside the agent (as explained before in the slack channel).

fixed

Jgmedina95:
Done!

SamCox822:
@Jgmedina95 my only concern now is that I don't think we need/should have pdb files for water & urea downloaded & saved in our repo. I don't think its particularly useful

Jgmedina95:
Agree, it is unnecessary now.

Jgmedina95 (approved):

PR #82, opened 2024-02-06
updated SkillRetrieval and ModifyScript descriptions and names

qcampbel:
I got a report where SkillRetrieval tool is chosen over ModifyScript tool, when the goal is to actually modify simulation scripts (just 5 lines of code). 

Looks like mdagent 'misunderstands' what both tools are for, so I opened this PR to address this. Changing tool descriptions may be all it takes, but if it doesn't, we may consider renaming tools as well.


qcampbel:
@Jgmedina95 when you have time, can you test this with your prompts for RDF procedure? 
While it doesn't contain packmol changes yet (other PR), the goal is that it chooses ModifySimulationScript more frequently now.

PR #84, opened 2024-02-08
80-adding-solvate-option-in-setupandrun-tool

Jgmedina95:
As part of my action items from yesterday meeting. 

Im adding the option to solvate the protein directly using modeller.addSolvent. This way, the agent can do a simulation directly without using packmol to just add water to the simulation. 

The prompt for testing was: 
""Simulate 1A3N in water at 280K. Dont forget to clean the protein first"

The second part is just to make it work quicker. The agent still needs the warning from the tools to then clean the file most of the time.


qcampbel (approved):
LGTM. Consider commenting out the old unused code

PR #85, opened 2024-02-14
Learn flag

SamCox822:
I added a learn flag to mdagent. This is so that if you want to force mdagent to only use existing/learned skills, rather than learn new skills, you can. I also added the flag to streamlit & added a unit test for initializing mdagent with the flag.

### DIFF
--- a/mdagent/mainagent/agent.py
+++ b/mdagent/mainagent/agent.py
@@ -70,16 +70,15 @@ class MDAgent:
             callbacks=[StreamingStdOutCallbackHandler()],
         )
 
-        # assign prompt
         if learn:
-            if agent_type == "Structured":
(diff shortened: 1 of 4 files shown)


Jgmedina95:
If im getting this right, the learn flag lets you use the Structure agent, but without it you can only use OpenAIfxns?

SamCox822:
> I concur with Jorge's comment. Since it's related to learning, can we also add confirmation message with Yes/No option whenever mdagent chooses to use CreateNewTool?

If I understand this, you want the system to ask the user every single time it goes to create a new tool? I don't think we want that

Jgmedina95 (approved):
LGTM!, make sure to pull main first. I merged a PR earlier

PR #91, opened 2024-02-21
Pqa tool

SamCox822:


mdagent/tools/base_tools/util_tools/search_tools.py
```diff
@@ -1,26 +1,75 @@
...
+    return answer
 
 
 class Scholar2ResultLLM(BaseTool):
     name = "LiteratureSearch"
-    description = """Input a specific question,
-                returns an answer from literature search."""
+    description = (
+        "Useful to answer questions that require technical "
+        "knowledge. Ask a specific question."
```
Jgmedina95, line 62: maybe add a couple examples of "Specific questions"

mdagent/utils/general_utils.py
Jgmedina95, line 1: where is this being used?

Jgmedina95 (commented):
Overall, i like the idea that we don't need a pqa API key anymore. My only question, not even a concern, is where is the function find_file_path used?

mdagent/utils/general_utils.py
SamCox822: thanks for point this out - its an artifact of something I was doing for another PR - I removed it

SamCox822:
> Overall, i like the idea that we don't need a pqa API key anymore. My only question, not even a concern, is where is the function find_file_path used?

fixed this - thank you!

Jgmedina95 (approved):
LGTM

PR #94, opened 2024-02-24
Registry

SamCox822:
This PR updates the tools to correctly use path registry and fixes minor bugs


mdagent/tools/base_tools/analysis_tools/plot_tools.py
```diff
@@ -8,60 +8,88 @@
...
+    ):
+        self.path_registry = path_registry
+        self.data = None
+        self.headers = None
+        self.matched_headers = None
+        self.file_name = None
+        self.file_path = None
+
+    def _find_file(self, file_name: str) -> None:
+        self.file_name = file_name
```
Jgmedina95, line 24: it would be self.file_id = file_id,

mdagent/tools/base_tools/analysis_tools/plot_tools.py
```diff
@@ -71,22 +99,29 @@ class SimulationOutputFigures(BaseTool):
...
+            plot_result = plotting_tools.plot_data()
             if type(plot_result) == str:
                 return "Figures created: " + plot_result
             else:
                 return "No figures created."
         except ValueError:
-            return "No timestep data found in csv file."
+            return "File could not be processed."
         except FileNotFoundError:
             return "Issue with CSV file, file not found."
```
Jgmedina95, line 126: Here I usually describe the problem stating that the name/id is not in the path_registry. check again

mdagent/tools/base_tools/analysis_tools/rmsd_tools.py
```diff
@@ -257,13 +277,17 @@ def _run(
...
+                self.path_registry, pdb_file, trajectory, ref_file, ref_trajectory
+            )
             message = rmsd.calculate_rmsd(rmsd_type, selection, plot)
         except ValueError as e:
             return (
                 f"ValueError: {e}. \nMake sure to provide valid PBD "
                 "file and binding site using MDAnalysis selection syntax."
             )
+        except FileNotFoundError as e:
+            return str(e)
```
Jgmedina95, line 290: yeah, i like to make this messages to the agent more descriptive.

Jgmedina95 (commented):
overall I like the new structure. Feels way more organized :). I had a couple comments, but I don't think they are critic, except for the ones regarding the files ids for plots. We can either continue with the ID we have so far for the same simulations: REC0,1,2_simid, or make a new one. I would think, that just keeping what we have, and making good descriptions should work :)

mdagent/tools/base_tools/analysis_tools/plot_tools.py
SamCox822: I don't think so - I decided to have it saved as file_name and don't use file_id anywhere in this class or file.

SamCox822: are you just asking me to rename it?

Jgmedina95: yeah, just renaming it haha. Although, we should keep it consistent through the entire package. I call file name at the literal file name "1XYZ.pdb" for example

Jgmedina95: Not saying im perfect, im almost certain I overlooked this somewhere

mdagent/tools/base_tools/analysis_tools/rmsd_tools.py
```diff
@@ -123,6 +123,9 @@ def compute_rmsd(self, selection="backbone", plot=True):
             plt.savefig(f"{self.filename}.png")
             # plt.close() # if you don't want to show the plot in notebooks
             message += f"Plotted RMSD over time. Saved to {self.filename}.png.\n"
+            self.path_registry.map_path(
+                f"{self.filename}.png", f"{self.filename}.png", message
```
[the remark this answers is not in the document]
SamCox822: what do you recommend here?

mdagent/tools/base_tools/analysis_tools/rmsd_tools.py
SamCox822: I add the description above where I define the filenotfound error message

SamCox822:
> overall I like the new structure. Feels way more organized :). I had a couple comments, but I don't think they are critic, except for the ones regarding the files ids for plots. We can either continue with the ID we have so far for the same simulations: REC0,1,2_simid, or make a new one. I would think, that just keeping what we have, and making good descriptions should work :)

most of the description errors you are seeing is because I describe the error when defining it, not when returning it. We can chat soon about the file id

mdagent/tools/base_tools/analysis_tools/rmsd_tools.py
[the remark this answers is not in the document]
Jgmedina95: So far ive been going with a "simple" approach. Anything that comes as an output from the simulation is a record, everything that ends in .pdb is a pdb file, and simulations are saved as scripts. 
Also if a file is a "child" of another file, ill just use that as part of the ID to keep it "traceable".

In this case, the plot comes from a Traj file with its time_stamp, so ill just give it an ID like it: REC0_TRAJ123456, where the number is the time stamp from the traj file.

Jgmedina95 (approved):
LGTM

PR #98, opened 2024-02-28
Feature rdftool

Jgmedina95:
Results of RDF testing


SamCox822 (changes requested):
Is it possible to combine these into 1 notebook instead of adding so many files? this pr edits 97 files

mdagent/tools/base_tools/analysis_tools/rdf_tool.py
```diff
@@ -0,0 +1,203 @@
...
+        "it will only calculate RDF for protein and water molecules.",
+    )
+    # atom_indices: Optional[List[int]] = Field(
+    #    None, description="Atom indices to load in the trajectory"
+    # )
+    # TODO: Add pairs of atoms to calculate RDF within the tool
+    ##pairs: Optional[str] = Field(None, description="Pairs of atoms to calculate RDF ")
+
+
+class RDFutils:
```
SamCox822, line 30: this doesn't appear to be used anywhere - can we remove unused code and extra comments?

mdagent/tools/base_tools/analysis_tools/rdf_tool.py
```diff
@@ -0,0 +1,203 @@
...
+        "stride: 2 \n"
+        "selections: None\n"
+        "Input Example 2: \n"
+        "trajectory_fileid: 'rec0_142404' \n"
+        "topology_fileid: 'top_sim0_142401' \n"
+        "\n\n"
+        "As you can see, the stride and selections are optional. "
+    )
+    args_schema = RDFToolInput
+    path_registry: Optional[PathRegistry]
```
SamCox822, line 51: you need to also have an init function for the path registry

mdagent/tools/base_tools/analysis_tools/rdf_tool.py
```diff
@@ -0,0 +1,203 @@
...
+            r, gr = md.compute_rdf(
+                traj,
+                traj.topology.select_pairs(
+                    ("protein and backbone and " "(name C or name N or name CA)"),
+                    "water and name O",
+                ),
+                r_range=(0.1, 2),  # Adjust these values based on your system
+                bin_width=0.005,
+            )
+        except Exception as e:
```
SamCox822, line 92: can you return the error instead of printing + raising another error?

SamCox822 (changes requested):

mdagent/tools/base_tools/analysis_tools/rdf_tool.py
Jgmedina95: So, the idea of raising it was to actually catch errors for fixing and handling them as they come. So if a ValueError came, i would then handle that specific error with their specific return statemetns

                bin_width=0.005,
```

File: mdagent/tools/base_tools/analysis_tools/rdf_tool.py [hunk 4]
```diff
@@ -0,0 +1,201 @@
...
+    )
+
+    topology_fileid: Optional[str] = Field(None, description="Topology file")
+    stride: Optional[int] = Field(None, description="Stride for reading trajectory")
+    selections: Optional[List[List[str]]] = Field(
+        [["protein", "water"]],
+        description="Selections for RDF. Do not use for now. As "
+        "it will only calculate RDF for protein and water molecules.",
+    )
+    # atom_indices: Optional[List[int]] = Field(
```
SamCox822, line 23: can you remove this?

SamCox822 (approved):

PR #99, opened 2024-02-28
Writing script unit tests were missing

SamCox822:


Jgmedina95:
LGTM!, one question? Does taking the simulation construction methods setup_system, etc... outside the init gives extra value? or is just better practice?

PR #101, opened 2024-03-01
Unit tests 

Jgmedina95:
Hello, @SamCox822 and @qcampbel. (whoever is reviewing haha).
So, as always, went a little over the board. I added the conftest.py file because our test file was getting too damm long. 

I made it such that the path registry, and some test files were centralized.
@SamCox822 , I THINK I went a little into what you did for testing the setup and run functionalities, but is fine. we don't have enough testing either way.


tests/test_fxns.py
```diff
@@ -44,39 +43,39 @@ def fibronectin():
     return "fibronectin pdb"
 
 
-@pytest.fixture
-def get_registry():
-    return PathRegistry()
+# @pytest.fixture
```
SamCox822, line 46: can you remove these, as they appear unnecessary here? (and everywhere else this is)

tests/test_fxns.py
```diff
@@ -160,13 +159,15 @@ def test_find_png(vis_fxns):
...
-def test_create_notebook(path_to_cif, vis_fxns):
+def test_create_notebook(ALA_CIF_file, vis_fxns):
+    path_to_cif = ALA_CIF_file
     result = vis_fxns.create_notebook(path_to_cif)
     path_to_notebook = path_to_cif.split(".")[0] + "_vis.ipynb"
     os.remove(path_to_notebook)
     assert result == "Visualization Complete"
 
 
+@pytest.mark.skip(reason="This has been deprecated")
```
SamCox822, line 170: Why are we skipping these now?

tests/test_fxns.py
```diff
@@ -425,14 +449,15 @@ def mock_listdir(path):
...
+def path_registry_with_mocked_fs(get_registry):
     with patch("os.path.exists", side_effect=mock_exists):
         with patch("os.listdir", side_effect=mock_listdir):
-            registry = PathRegistry()
+            registry = get_registry("raw", False)
             registry.get_timestamp = lambda: "20240109"
             return registry
 
 
+@pytest.mark.skip(reason="This was moved to test_pathregistry.py")
```
SamCox822, line 460: If this is the case, can you remove it here, rather than just skipping?

SamCox822 (changes requested):
Thanks for adding these! Quick notes:
- rather than comment things out, pls just remove it (as it appears to be redundant in most of these cases)
- can we avoid skipping tests for now? if they need fixed, we can fix them & if we need to remove them, we should
- also can you fix the merge conflicts?

thanks!

tests/test_fxns.py
Jgmedina95: sure

tests/test_fxns.py
Jgmedina95: We dont use the SimulationFunctions class anymore since we move to multi-input tools :)

SamCox822: then can we remove it>

Jgmedina95: yes we can

tests/test_pathregistry.py
```diff
@@ -0,0 +1,42 @@
+import os
+
+import pytest
+
+
+@pytest.mark.parametrize("with_files, raw_or_clean", [("raw", False), ("raw", True)])
+def test_registry_init(get_registry, with_files, raw_or_clean):
+    # make the test directory the cwd
```
SamCox822, line 8: can you remove this? thanks!

tests/test_setupandrun.py
```diff
@@ -0,0 +1,115 @@
...
+            }}
+        }}
+        """.format(
+            pdb_id=pdb_id
+        ).strip()
+
+    return create_input
+
+
+# @pytest.fixture(scope="module")
```
SamCox822, line 64: can you remove this too?

SamCox822 (approved):
it looks good to me - can you just remove the extra comments before merging in? thanks!

PR #102, opened 2024-03-04
add eval for each step

qcampbel:
implementing automated evaluation for each step of MDAgent


qcampbel:
@SamCox822 @Jgmedina95 can you take a look at `run_and_eval` function and let me what are other metrics I need to add to this evaluation? Thx!

Jgmedina95:
This looks great! I wonder if you could also add a little example of how it works? nothing fancy, a bunch of prints in between would be enough :)

qcampbel:
> This looks great! I wonder if you could also add a little example of how it works? nothing fancy, a bunch of prints in between would be enough :)

Just added an example. Also added more eval info such as agent type, model name, etc.

Jgmedina95 (approved):
Nice!, any reason why you think all the numbers in the examples have like 10 decimals even though in two of them you specified 3 decimals? Is not deal braking, but it would be nice if all of them had only up to 3 decimal too :)

qcampbel:
@SamCox822 @Jgmedina95 please take a look at notebooks/eval_demo and let me know what you think!

Jgmedina95:
Looks pretty solid to me!

PR #104, opened 2024-03-06
Clean tools tests

SamCox822:


Jgmedina95 (commented):
Hi! The problem here is that the cleaning tools, do not use this utilities anymore, they were embedded directly into the tool when we transitioned to multi-input functions. 
Now the cleaning is all done by the Fixer. Like it does all those things in a test like this
```def test_cleaning_function(get_registry):
    reg = get_registry("raw", True)
    tool = CleaningToolFunction(path_registry=reg)
    assert tool.name == "CleaningToolFunction"
    assert tool.path_registry == reg
    print(reg.get_mapped_path("ALA_123456"))
    print(os.listdir("."))
    print(os.listdir("files/pdb/."))
    prompt = {
        "pdb_id": "ALA_123456",
        "replace_nonstandard_residues": True,
        "add_missing_atoms": True,
        "remove_heterogens": True,
        "remove_water": True,
        "add_hydrogens": True,
        "add_hydrogens_ph": 7.0,
    }
    result = tool._run(**prompt)
    assert "File cleaned" in result
```

SamCox822:
So these functionalities are no longer used at all or no longer used individually? we need to either remove them or test them individually like this, no? @Jgmedina95

PR #105, opened 2024-03-06
Packmol tests

SamCox822:


Jgmedina95 (dismissed):

Jgmedina95:
Adding some tests on catching errors at the instructions. Example: Only one instruction, but two Molecules where passed. And vice-versa. Also, packmol handles different types of instructions. like centerofmass, center, etc... Some of them are not valid and are included in the validation of the inputs in the tool, so checking if this work correctly could be added. 
A good one is if the validation can check if the pdb ids, are actually currently in the path registry to move forward or not. 

FOr inspiration on the last item, in my PRs I did some of this :)

Jgmedina95:
I see what you're doing here, and I like it! Need help fixing the errors at testing?

Jgmedina95 (approved):
LGTM!

Issue #97, opened 2024-02-27
Add FIGURES in the path_registry 


Jgmedina95:
RMSD tool (And other properties) output, will be a plot, so it makes sense to add a FIGURE or PLOT file type in the registry. Thanks @SamCox822 for the idea.

PR #108, opened 2024-03-12
Figures path registry

Jgmedina95:
As mentioned in issue #97. Figures are been added to the path registry. They will be saved into the files/figures directory.
So far I've made the appropriate changes to plot tools, rmds tools and rgy tools. As those are the ones in main as of now.

I still need to test, and add unit testing at the path registry for the new functionalities so don't panic @SamCox822


SamCox822 (approved):
lgtm

PR #111, opened 2024-03-19
Checkpoints

SamCox822:


mdagent/utils/set_ckpt.py
```diff
@@ -0,0 +1,62 @@
+import os
+import shutil
+
+
+class SetCheckpoint:
+    def find_root_dir(self):
```
Jgmedina95, line 6: i have a proposition: 
```
def find_project_root(stop_at="md-agent"):
    """
    Traverse up from the current_path to find the stop_at directory.

    :param current_path: The starting point for the search.
    :param stop_at: The directory name to stop the search at.
    :return: The path to the directory named stop_at, or just current_path if not in
    a directory named stop_at or any of its parents.
    """
    current_path = Path(os.getcwd())
    if current_path.name == stop_at:
        return current_path
    for parent in current_path.parents:
        if parent.name == stop_at:
            return parent
    return current_path
```
I had this so that the ckpt will always go to the root directory of md-agents, or in case we're not working inside the repo, it will just be the local directory.

mdagent/subagents/subagent_setup.py
```diff
@@ -12,17 +12,18 @@ def __init__(
         temp=0.1,
         max_iterations=40,
         verbose=True,
-        ckpt_dir="ckpt",
         resume=False,
         retrieval_top_k=5,
         curriculum=True,
     ):
         self.path_registry = path_registry
+        if self.path_registry is None:
```
Jgmedina95, line 20: how about raising an error instead? it would be problematic if the sub agents have a different checkpoint, than the other tools right?

mdagent/utils/path_registry.py
```diff
@@ -16,17 +18,46 @@ class FileType(Enum):
...
+        if resume:
+            self.ckpt_dir = self.set_ckpt.get_resume_ckpt()
+            if self.ckpt_dir is None:
+                self.ckpt_dir = self.set_ckpt.set_ckpt_subdir(ckpt_dir=ckpt)
+        else:
+            self.ckpt_dir = self.set_ckpt.set_ckpt_subdir(ckpt_dir=ckpt)
+        return None
+
+    def _make_all_dirs(self):
+        self.json_file_path = os.path.join(self.ckpt_dir, "paths_registry.json")
```
Jgmedina95, line 44: love this btw

Jgmedina95 (commented):
just some minor comments that I would like to talk before approving :)

mdagent/subagents/subagent_setup.py
SamCox822: AH, get instance gets the current ckpt, it doesn't make a new one :)

mdagent/utils/path_registry.py
SamCox822: ty :)

Jgmedina95 (approved):
LGTM

Issue #112, opened 2024-03-19
Bug at openMMSimulation object


Jgmedina95:
as @brittyscience pointed out yesterday. There was a bug during the simulation initialization. 

_Bug description:_
when getting the simulation parameters this was used:
self.params.integrator_params, (I believe this was done by @SamCox822 in one of the many PRs we did this last week)

The issue was solved just using keys instead of attributes
self.params.integrator_params --> self.params["integrator_params"] 

This works! Thanks @brittyscience for pointing the bug.

PR #113, opened 2024-03-19
Changing attribute to key in params input in openmm simulation

Jgmedina95:
Fixing issue #112


Jgmedina95:
The issue was created, i believe, because of confusion between the FunctionInputs object, and the actual input to the tools. 
The actual input to the tools is just a string, that gets transformed into a dictionary. While the FunctionInputs object is used as a guide for the model. 

If you use the FunctionInputs (like it was during testing) as the input to OpenMMSimulation, it would work, but the agent doesn't work like that, and the input is just a dictionary (obviously following the Inputs guideline) so it doesn't have attributes, it has keys.

qcampbel (approved):

PR #118, opened 2024-03-27
Memory

SamCox822:


Jgmedina95:
I think this is great! Although, i tested if it was able to continue from one step without repeating it. Downloading a file, and then preprocess it, but seems to not be the case.

Do you think it can be done?
Also, when you ran the agent for the second time, the memories get replaced, not appended. Is that intentional to avoid long context, or a bug?

Jgmedina95:
Hi! was looking over here, and came up with a suggestion:
in the memory, to avoid reinitializing the memories: how about just putting the summaries in a list? It can be easier to keep adding extra summaries that way :) 
```
    def _write_to_json(self, data, path):
        key = data.keys()[0]
        summary = data[key]
        if os.path.exists(path):
            data_original = json.load(open(path))
            if key in data_original.keys(): 
                if len(data_original[data.keys()[0]]) > 0:
                    data_original[data.keys()[0]].append(summary)
                else:
                    data_original[data.keys()[0]] = [summary]
            else:
                data_original.update(data)
        else:
            data_original = {key: [summary]}
        with open(path, "w") as f:
            f.write(json.dumps(data_original))
```

Jgmedina95 (approved):
LGTM! The only change i would suggest is modifying to write the json file:
```
f.write(json.dumps(data_original, indent=4, sort_keys=True))
```
So it looks pretty when you look read the memory.

PR #123, opened 2024-04-01
Prompting change in CNT

Jgmedina95:
Hi all, here is the branch with the changes I made in the CNT prompts :) Because I needed some new arguments (like args in CNT), I had to change some tools at their init too.


qcampbel:
just added minor bug fix that we discussed before about CreateNewTool failing to get expected "args" argument. Looks good to me!

qcampbel (approved):

PR #124, opened 2024-04-22
adding argument validator

Jgmedina95:
Im adding a new argument validator to check whenever the agent makes either a typo mistake or just a wrong argument name:
e.g. pdb = ABC instead of pdb_file = ABC. 
The argument validator checks the arguments and gives back a suggestion on the argument (more descriptive error message)

Example from a test run:
```
Action:
{
  "action": "RDFTool",
  "action_input": {
    "input": {
      "topology": "top_sim0_112831",
      "trajectory": "rec0_112833"
    }
  }
}
```
The answer from the tool is:
```
Invalid Arguments in RDF tool:  Invalid argument(s) provided. 'topology': (maybe you mean: 'topology_fileid'?)
 'trajectory': (maybe you mean: 'trajectory_fileid'?)
```
Which makes it easier for the agent to correct itself. 
This is better than just having a general error message like before ("The inputs are wrong, try again with the correct arguments")


Jgmedina95:
@SamCox822 , or @qcampbel. I think this is going to be helpful, an example from one of my runs:
```
Action:
```json
{
  "action": "RDFTool",
  "action_input": {
    "input": {
      "trajectory_file": "rec0_141659",
      "topology_file": "top_sim0_141657",
      "output_file": "rdf_1A3N_water",
      "rdf_type": "protein-water"
    }
  }
}
```Invalid Tool argument(s) provided: trajectory_file: Did you mean 'trajectory_fileid'?, topology_file: Did you mean 'topology_fileid'?, output_file: This argument is not recognized and will be ignored., rdf_type: This argument is not recognized and will be ignored.
The RDFTool requires the correct argument names to be used. The error messages indicate that the correct argument names are 'trajectory_fileid' and 'topology_fileid', and that 'output_file' and 'rdf_type' are not recognized arguments. I need to correct the argument names and remove the unrecognized arguments before retrying the RDF calculation.

Action:
```json
{
  "action": "RDFTool",
  "action_input": {
    "input": {
      "trajectory_fileid": "rec0_141659",
      "topology_fileid": "top_sim0_141657"
    }
  }
}
```Names found in registry: 1A3N_141604, 1A3N_141618, top_sim0_141657, sim0_141657, rec0_141659, rec1_141659
The Radial Distribution Function (RDF) of the protein 1A3N with respect to water molecules has been successfully calculated and the resulting graph has been saved as 'rdf.png' in the current directory. The task is now complete.
```

As you can see, the tool directly answers back with an error message indicating what argument was wrong, and then suggests corrections. Particularly useful in longer runs where the agent for some reason becomes idiotic, and fails at proposing the correct arguments. 

Note: This only checks for argument correctness (like if the argument is "file_id" instead of "file") it doesn't check the value. 

Note 2: Ive also added a very similar function in the utils/validator that works similarly but for functions (this will come handy with the Create New Tool functions too!)

qcampbel:
really like the implementation - make it so simple to validate arguments. however, I don't see `validate_func_args` being used? what do you plan to use it for? so far, we just need to validate tool args for MD-Agent to troubleshoot.

Jgmedina95:
Right now when the agent makes an argument mistake, the error messages are non-specific ("arguments are wrong, correct them, remember this tool needs A,B, etc..."). With this, if the arguments are wrong (Which tend to happen in longer runs) The validator automatically answers with which argument exactly was wrong and what was expected instead. Just makes it easier for the agent to correct itself.

Jgmedina95:
oh! and the validate_func_args, could possibly be used with new tools written by CNT. As of now, they are mainly functions right? So that's why I add it.

SamCox822:
I don't think we should implement this, as it seems to be overfitting to specific use cases. If we can think of a more general approach, then maybe worth revisiting.

SamCox822:
@Jgmedina95 can we close this? I dont think its relevant anymore

SamCox822:
closing

PR #128, opened 2024-05-07
scrap CreateNewTool

qcampbel:
removed CreateNewTool (CNT) from the main codebase and moved the current codebase with this CNT feature under `dev-createnewtool` branch


qcampbel:
A couple of things I haven't addressed here:

PR #129, opened 2024-05-09
Debugging: I think I tackled here 8 or so issues/bugs of our internal doc

Jgmedina95:


mdagent/utils/path_registry.py
```diff
@@ -306,11 +305,26 @@ def write_file_name(self, type: FileType, **kwargs):
                 file_name += f"{type_of_sim}_{protein_file_id}_{time_stamp}.py"
         if type == FileType.RECORD:
             record_type_name = kwargs.get("record_type", "RECORD")
-            term = kwargs.get("term", "term")  # Default term if not provided
+            if Sim_id and protein_file_id:
+                file_name = (
```
SamCox822, line 309: Aren't these checks the same (lines 308,313)? Can we condense this entire chunk?

SamCox822 (changes requested):

mdagent/utils/path_registry.py
Jgmedina95: Got it! Accidentally double it...

At first I thought you meant the whole function, which I Know can be better. I could separate each separate case in subfunction to handle each type of file to improve readability. (_write_protein_name, _write_record_name...) and even make a generator to make it fancier. Would you like that in this PR?

SamCox822: I don't think it needs to be fancier, I'm wondering if it can be simpler lol

SamCox822 (approved):

PR #130, opened 2024-05-15
Secondary Structure & Shape Tools from MDTraj

SamCox822:
tools created, still need to pass to agent and write unit tests


mdagent/tools/base_tools/analysis_tools/secondary_structure.py
```diff
@@ -0,0 +1,301 @@
...
+    xtc, .trr) and an optional topology file (e.g., .pdb, .prmtop). The
+    output is an array of acylindricity values for each frame of the
+    trajectory."""
+    path_registry: PathRegistry | None = None
+
+    def __init__(self, path_registry: PathRegistry):
+        super().__init__()
+        self.path_registry = path_registry
+
+    def _compute_acylindricity(self, traj):
```
Jgmedina95, line 222: It would be nice to include a little doc string on what is the output of the function.

Jgmedina95: the same in the other tools

mdagent/tools/base_tools/analysis_tools/secondary_structure.py
Jgmedina95, line 1: I remember something similar in Quinny's PR from a few weeks ago. This functions take a traj file as input, can they also take a pdb file and do it over only one frame? Cause in that case, the result may just be a single value that could be reported instead of being plotted. Or depending on what is reported (don't think is applicable in this functions) a value per residue?

mdagent/tools/base_tools/analysis_tools/secondary_structure.py
```diff
@@ -0,0 +1,301 @@
...
+        summary = self._summarize_dssp(dssp_array)
+        return summary
+
+    async def _arun(self, traj_file, top_file=None):
+        raise NotImplementedError("Async version not implemented")
+
+
+class ComputeGyrationTensor(BaseTool):
+    name = "ComputeGyrationTensor"
+    description = """Compute the gyration tensor for each frame in a
```
Jgmedina95, line 103: In the other tool descriptions I try to include trajectory file ID , instead of traj file (.xtc, .trr) as an old issue was the agent using the full path to get the file, instead of the file ID.

mdagent/tools/base_tools/analysis_tools/secondary_structure.py
Jgmedina95, line 1: After going through the documentation, the DSSP tools returns an (n_frames, n_residues) matrix with the code of which structure is which residue. 
The total count, makes sense, but maybe something like what I found in this paper can be cool to see: 
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7073106/. In particular, figure 4.

SamCox822: I see, yes that is the case. I can add a check & if there is only one frame, it doesn't make sense to plot it. 

some of the tools return an array though. I guess for now, we can return the array to the agent & see how it handles it. Or do you still think we should write it to a file for now if it's one frame? @Jgmedina95

SamCox822: it would be cool, but I think maybe out of scope for this project?

SamCox822:
The failing test seems unrelated to this PR @Jgmedina95

Jgmedina95:
> The failing test seems unrelated to this PR @Jgmedina95

Very very strange behavior. 
https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.integrate.simps.html

Im trying to find what it is about it, but I would agree, this has nothing to do with your PR

Jgmedina95:
solving with PR #144

mdagent/tools/base_tools/analysis_tools/secondary_structure.py
```diff
@@ -0,0 +1,551 @@
...
+        )
+
+    async def _arun(self, traj_file, top_file):
+        """Runs the tool asynchronously."""
+        raise NotImplementedError("Async version not implemented")
+
+
+class AnalyzeProteinStructure(BaseTool):
+    name = "AnalyzeProtein"
+    description = (
```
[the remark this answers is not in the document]
Jgmedina95: cause as of now, if instead of n_chains, the agent can use for example 'total chains', if he actually knows how haha, as it doesn't say what he can ask for.

mdagent/tools/base_tools/analysis_tools/secondary_structure.py
```diff
@@ -0,0 +1,551 @@
...
+            result["time"] = traj.time
+        if "time_step" in requested_analyses:
+            result["time_step"] = traj.time_step
+        if "atoms" in requested_analyses:
+            result["atoms"] = traj.topology.atoms
+        if "bonds" in requested_analyses:
+            result["bonds"] = traj.topology.bonds
+        if "chains" in requested_analyses:
+            result["chains"] = traj.topology.chains
+        if "residues" in requested_analyses:
```
Jgmedina95, line 532: do we want the total list of residues? or the list of unique residues? the total list of residues can be veery extent, and of atoms, even longer. So im not sure if is a good idea to return all of them. Maybe a summary of the residues would make more sense like: {"ALA": total count 35....}. with atoms, even less sure.

Now that im looking at the documentation, this tool won't work. 
atoms, residues, chains, and bonds are generators, so they can be used in a for loop, but not to see the list as strings. 

```
print(str(traj.topology.residues))
>> <generator object Topology.residues at ....> 
```

Here is some code that I was experimenting on:
```
new_topology = pd.DataFrame(
    {'serial':range(traj.n_residues),
    'name':['COM' for _ in traj.topology.residues],
    'resSeq':range(traj.n_residues),
    'resName':[i.name for i in traj.topology.residues],
    'element':["VS" for _ in traj.topology.residues],
    'chainID':[i.chain.chain_id if i.chain.chain_id else 0 for i in traj.topology.residues],
    'segmentID':[i.segment_id for i in traj.topology.residues]
    }
)
print(new_topology.head())
>>  serial name  resSeq resName element  chainID segmentID
0       0  COM       0     MET      VS        0      4AKE
1       1  COM       1     ARG      VS        0      4AKE
2       2  COM       2     ILE      VS        0      4AKE
3       3  COM       3     ILE      VS        0      4AKE
4       4  COM       4     LEU      VS        0      4AKE
```

Not that this is what you wanted, but to show how to work with the generators :) Again, instead of huge lists, a summary of the number of residues, atoms, and their ids per chain would be nicer?

mdagent/tools/base_tools/analysis_tools/secondary_structure.py
```diff
@@ -0,0 +1,551 @@
...
+            return (
+                "Gyration tensor computed for "
+                "a single frame, no file saved."
+                f"Gyrations tensor: {gyration_tensors}"
+            )
+
+        file_id = write_raw_x(
+            "gyration_tensor", gyration_tensors, traj_file, self.path_registry
+        )
+        return "Gyration tensor computed successfully, " f"saved to {file_id}"
```
Jgmedina95, line 224:
```
f"Gyration tensor computed successfully, saved to {file_id}"
```

mdagent/tools/base_tools/analysis_tools/secondary_structure.py
```diff
@@ -0,0 +1,551 @@
...
+        raise NotImplementedError("Async version not implemented")
+
+
+class ComputeAcylindricity(BaseTool):
+    name = "ComputeAcylindricity"
+    description = """Compute the acylindricity for each frame in a
+    molecular dynamics trajectory.
+    Input is a trajectory file ID and an optional topology file ID.
+    The output is an array of acylindricity values for
+    each frame of thetrajectory."""
```
Jgmedina95, line 347: ```the trajectory```

tests/test_analysis/test_secondary_structure.py
```diff
@@ -0,0 +1,154 @@
...
+    compute_relative_shape_antisotropy = ComputeRelativeShapeAntisotropy(
+        path_registry=registry
+    )
+    output = compute_relative_shape_antisotropy._compute_relative_shape_antisotropy(
+        loaded_cif_traj
+    )
+    assert np.allclose(output, np.array([0.26852832]))
+
+
+def test_analyze_protein_structure(get_registry, loaded_cif_traj):
```
Jgmedina95, line 145: yeah, depending on how the tool changes, the test would need to update as well :)

mdagent/tools/base_tools/analysis_tools/secondary_structure.py
SamCox822: lol - you're right

I think this tool really just makes sense to count, I dont think there is much value otherwise

mdagent/tools/base_tools/analysis_tools/secondary_structure.py
SamCox822: fixed

mdagent/tools/base_tools/analysis_tools/secondary_structure.py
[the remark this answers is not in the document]
SamCox822: fixed - I changed the fan to always give a count of all

Jgmedina95 (approved):
LGTM!

PR #131, opened 2024-05-15
thermo analysis tools

SamCox822:


mdagent/tools/base_tools/thermo_tools.py
```diff
@@ -0,0 +1,198 @@
...
+        raise NotImplementedError("Async version not implemented")
+
+
+class ComputeDensity(BaseTool):
+    name = "ComputeDensity"
+    description = """Calculate the mass density of each frame in a
+    trajectory. Requires a trajectory file and an array of masses and
+    optionally a topology file. Returns an array of mass densities for
+    each frame."""
+    # todo -> should this write to a file or return the array?
```
SamCox822, line 174: since its each frame

- plot it over the frames
- write it to a file (source data for the plot is accessible
- return to the agent some message about it

mdagent/tools/base_tools/thermo_tools.py
```diff
@@ -0,0 +1,198 @@
...
+    return traj if traj else None
+
+
+class ComputeDipoleMoments(BaseTool):
+    name = "ComputeDipoleMoments"
+    description = """Compute the total dipole moment for each frame in a
+    molecular dynamics trajectory. Requires a trajectory file and
+    optionally a topology file and partial charges. Returns an array of
+    dipole moments for each frame of the trajectory."""
+    # TODO -> should this write to a file or return the array?
```
SamCox822, line 22: same as other comment -> plot

mdagent/tools/base_tools/analysis_tools/thermo_tools.py
```diff
@@ -0,0 +1,335 @@
...
+        self, traj_file: str, top_file: str | None = None, charges: np.ndarray = None
+    ):
+        raise NotImplementedError("Async version not implemented")
+
+
+class ComputeStaticDielectric(BaseTool):
+    name = "ComputeStaticDielectric"
+    description = """Compute the static dielectric constant of a system
+      from the dipole moments of a molecular dynamics trajectory.
+      Requires a trajectory file, charges, temperature and optionally a
```
Jgmedina95, line 134: trajectory file ID

Jgmedina95: Does the agent knows what arguments to use without describing them? particularly the top_file, which is the only kwarg

mdagent/tools/base_tools/analysis_tools/thermo_tools.py
```diff
@@ -0,0 +1,335 @@
...
+                traj_fileid=traj_file,
+                top_fileid=top_file,
+            )
+            if not traj:
+                raise Exception("Failed to load trajectory file.")
+        except Exception as e:
+            return str(e)
+
+        try:
+            charges = get_charges(self.path_registry, traj, charge_file)
```
Jgmedina95, line 115: How is the agent getting or making the charge_file prior to using this tool?

Jgmedina95: i think in the case of a vector, it would make sense to also capture descriptors such as the magnitude and direction w.r.t. to the z-axis (?) or something. 

And even though I think I've said this a lot (sorry!) maybe the magnitude of the dipole per frame and or its direction can be cool figures

Jgmedina95 (commented):
yeah, my comments are above. Other than that, it seems fine! 
I notice you never really explicitly mention the arguments for the tools. In my tools, the args_schema do that implicitly, but I wonder if you consider it not necessary here too.

mdagent/tools/base_tools/analysis_tools/thermo_tools.py
SamCox822: I think I fixed this

mdagent/tools/base_tools/analysis_tools/thermo_tools.py
SamCox822: ah, the user can provide it. I added another function though that can try to load the traj into a system and get charges using openmm if not provided

mdagent/tools/base_tools/analysis_tools/thermo_tools.py
```diff
@@ -131,15 +194,21 @@ class ComputeStaticDielectric(BaseTool):
...
+      Returns the static dielectric
       constant."""
     path_registry: PathRegistry = PathRegistry().get_instance()
+    get_charges = GetTrajCharges()
 
     def __init__(self, path_registry: PathRegistry | None = None):
         super().__init__()
         if path_registry is not None:
             self.path_registry = path_registry
+        self.get_charges = GetTrajCharges()
```
Jgmedina95, line 211: is this one needed?

Jgmedina95 (approved):
Code wise, looks good to me. I didn't ran the tools by myself though. So just make sure that usual bugs are covered: (files saved with same name, or plots /filesnot closed, or saved to the registry)

SamCox822:
Closing this PR for now, as I'm not sure how useful these tools actually are. We can add pieces back later if needed.

PR #132, opened 2024-05-15
PCA Tool

Jgmedina95:


mdagent/tools/base_tools/analysis_tools/pca_tools.py
```diff
@@ -0,0 +1,370 @@
...
+    ):
+        self.path_registry = path_registry
+        self.traj = md.load(traj_path, top=top_path)
+        self.sim_id = sim_id
+        if pc_percentage > 1:
+            pc_percentage /= 100
+        self.pc_percentage = pc_percentage
+        self.atom_indices = self.traj.top.select(selection)
+        self.selection = selection
+        self.pc = PCA(n_components=30)
```
SamCox822, line 34: should n_components be a parameter?

mdagent/tools/base_tools/analysis_tools/pca_tools.py
```diff
@@ -0,0 +1,370 @@
...
+        ###########MD TRAJ ALIGNMENT################
+        try:
+            self.traj.superpose(self.traj, frame=0, atom_indices=self.atom_indices)
+            return "Trajectory aligned to the first frame. "
+        except Exception as e:
+            print(f"Error aligning trajectory: {str(e)}")
+
+    def get_pc(self):
+        align_message = self._align_trajectory()
+        # self.pc = pca.PCA(
```
SamCox822, line 49: can we get rid of the commented out pieces or do we need them for some reason?

mdagent/tools/base_tools/analysis_tools/pca_tools.py
```diff
@@ -0,0 +1,370 @@
...
+            (2.0 / T)
+            * (scipy.integrate.simps(cos * pca_space[:, i])) ** 2
+            / scipy.integrate.simps(pca_space[:, i] ** 2)
+        )
+
+    def make_scree_plot(self):
+        extra_mess = ""
+        if not self.pc:
+            pc_mess = self.get_pc()
+            if "not aligned" in pc_mess:
```
SamCox822, line 120: what is the purpose of this??

mdagent/tools/base_tools/analysis_tools/pca_tools.py
```diff
@@ -0,0 +1,370 @@
...
+        desc = f"Scree Plot of the eigenvalues of simulation {self.sim_id}"
+        file_name = self.path_registry.write_file_name(
+            FileType.FIGURE,
+            fig_analysis="scree_plot",
+            file_format="png",
+            Sim_id=self.sim_id,
+        )
+        plot_id = self.path_registry.get_fileid(file_name, FileType.FIGURE)
+        path = self.path_registry.ckpt_dir + "/figures/"
+        plt.savefig(path + file_name)  # Save the figure before plt.show()
```
SamCox822, line 150:
```suggestion
        plt.savefig(path + file_name) 
```

This comment appears several times in this file, can we remove it? especially since we aren't actually showing the plot anywhere

mdagent/tools/base_tools/analysis_tools/pca_tools.py
```diff
@@ -0,0 +1,370 @@
...
+    )
+    topology_fileid: str = Field(
+        description=("Topology File ID of " "the simulation to be analyzed")
+    )
+    pc_percentage: Optional[float] = Field(
+        95.0, description="Max cumulative percentage of components for analysis"
+    )
+    analysis: str = Field(
+        "all",
+        description=(
```
SamCox822, line 217: Have you tested this functionality this way? I thought we decided that separate tools would be better than a combo tool with input params to route.. @qcampbel what do you think here?

SamCox822: If it works consistently better than the alternative, I'm open to changing it to this approach. I haven't seen this combo approach have higher performance in my experience though.? Whichever way we decide, I'd like to keep things consistent, so we might need to make other changes in the code to reflect it

mdagent/tools/base_tools/analysis_tools/pca_tools.py
```diff
@@ -0,0 +1,370 @@
...
+        "performs analysis with them"
+    )
+    args_schema = PCASchema
+    path_registry: Optional[PathRegistry]
+
+    def __init__(self, path_registry: Optional[PathRegistry] = None):
+        super().__init__()
+        self.path_registry = path_registry
+
+    def _run(self, input):
```
SamCox822, line 251: can we separate this out into multiple functions? it might be easier to debug & is probably cleaner to split it up

tests/test_pca_tool.py
```diff
@@ -0,0 +1,46 @@
...
+        "seaborn.PairGrid"
+    ), patch(
+        "seaborn.PairGrid.map"
+    ):
+        result = pca_tool._run(good_inputs)
+
+    assert "Analyses done:" in result
+    assert "Cosine Content of each PC: " in result
+
+    # This two tests involve matplotlib, and are annoying to get done.
```
SamCox822, line 44: Can you instead split them up to test them and add an ignore flag for pytest? Ideally, we should have 1 test for the bad inputs, 1 for the good, and then you can separate the matplotlib stuff into another function that we can tell pytest to ignore for now

SamCox822 (changes requested):
Thanks for adding this feature! I left a few comments, but additionally, it seems like there are a lot of helper functions / other functions that don't have unit tests. can you add those?

mdagent/tools/base_tools/analysis_tools/pca_tools.py
Jgmedina95: thanks! i removed a ton from the MDanalysis functions, but I guess some went unnoticed be me.

mdagent/tools/base_tools/analysis_tools/pca_tools.py
Jgmedina95: oki. Ive noticed this comes because of auto-completion with copilot is the root reason of these repetitive comments haha. I've deal with two of them so far, and if I keep seeing them, ill remove them too :)

mdagent/tools/base_tools/analysis_tools/pca_tools.py
Jgmedina95: maybe, I'm putting 30 here as default, as in some cases no more than 3-4 is necessary to carry >95% of the variance, and otherwise is an indication that the simulation didn't run long enough. 

Im putting 30, and then filtering out to only keep the top 95% of variation. If 30 is not enough... then either the simulation is just random, or has not converged yet (needs more time)

Jgmedina95: But this is up to agreement though, I decided on 30 based on some examples I saw while making the tool

mdagent/tools/base_tools/analysis_tools/pca_tools.py
Jgmedina95: Before making the PCA calculation, is important to align the trajectory (usually against the first frame), so it captures relative movement of atoms against themselves. But if is rotations, and or translations are not taking into account they can skew the analysis. Even then, the analysis can still be performed, but the results are less trustful, which is why I put that as part of the final message from the tool.

mdagent/tools/base_tools/analysis_tools/pca_tools.py
SamCox822: hm okay

mdagent/tools/base_tools/analysis_tools/pca_tools.py
SamCox822: Can we include it as a warning to the user, rather than a printed statement or message to the agent?

mdagent/tools/base_tools/analysis_tools/pca_tools.py
SamCox822: We'll test this in a separate PR.

mdagent/tools/base_tools/analysis_tools/pca_tools.py
Jgmedina95: This is a good idea

qcampbel:
Jorge, can you add "Succeeded. " or "Failed. " at the beginning of every message from tool? That way, we can still use evaluator code, to make our evaluations easier & faster.

mdagent/tools/base_tools/analysis_tools/pca_tools.py
Jgmedina95: done! split into get_values, and run_pca_analysis functions. it reads better now

Jgmedina95:
@qcampbel @SamCox822 , I think I addressed all the comments :) Please let me know if there's something else!

tests/test_pca_tool.py
SamCox822: can we do this ^^?

mdagent/tools/base_tools/analysis_tools/pca_tools.py
Jgmedina95: alright, sorry for the misunderstanding haha I added a warning now, instead of just printing out.

SamCox822 (approved):

SamCox822:
@Jgmedina95 can you put the tests in the appropriate folder in main before merging?

PR #133, opened 2024-05-15
added dummy trajectory files for testing

qcampbel:
- added trajectory files for testing
- add map_path option in `get_registry` fixture so paths to these files are mapped automatically
- updated unit test for path registry that is largely impacted


qcampbel:
@Jgmedina95 and @SamCox822 --- A potential caveat (?): in these trajectory files, all frames are identical to each other (not moving). Meaning, unit tests should give expected zero energy, RMSD, or whatever you're measuring related to time. 
If that's not what you want, I can expand the code to move slightly for each frame or make another quick trajectory file if the need for it arises.

PR #135, opened 2024-05-22
SASA & inertia tools

qcampbel:


Jgmedina95:
For the Moment of Inertia tool im in a bit of a loss. Does the tensor itself gives something useful? I was trying to see how people use it and found some examples where the Eigenvectors and eigenvalues of each tensor per frame are actually useful for some interpretation. 

https://sites.psu.edu/simtech/micelles/3/ shows a nice plot with the three eigenvalues for each frame and made an analysis on the eccentricity. 

Here https://doi.org/10.1002/advs.202304262 the authors analyzed the principal axes of rotation to see and calculation auto-correlation times with it. 

I don't know the meaning of the average of the tensor itself, but I would consider to use the tool to give back the eigenvalues per frame (and that can also be plotted).

Jgmedina95:
Hi @qcampbel is this ready for re-review?

qcampbel:
@Jgmedina95 I'd like to test it throughly right before merging just to double check everything. But sure, do leave a review! I've addressed all of your comments here

mdagent/tools/base_tools/analysis_tools/sasa.py
```diff
@@ -0,0 +1,156 @@
...
+from langchain.tools import BaseTool
+from pydantic import BaseModel, Field
+
+from mdagent.utils import FileType, PathRegistry
+
+
+class SASAAnalysis:
+    def __init__(self, path_registry, top_fileid, traj_fileid=None, mol_name=None):
+        """
+        Initialize the MomentOfInertia class with topology and/or trajectory files.
```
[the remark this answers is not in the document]
qcampbel: good catch!

mdagent/tools/base_tools/analysis_tools/sasa.py
```diff
@@ -0,0 +1,154 @@
...
+        """
+        message = ""
+        if self.sasa is None or self.residue_sasa is None:
+            message += self.calculate_sasa()
+
+        # Returns areasnp.array, shape=(n_frames, n_features)
+        # if there's only one frame, don't plot
+        if self.sasa.ndim == 1:
+            message += " Only one frame in trajectory. No SASA plot generated."
+            return message
```
Jgmedina95, line 87: we can also add that only value right?

```suggestion
            message += f"Available Surface area is: {self.residue_sasa}. 
            return message
```

Jgmedina95:
For the SASA tool i ran it to test it and got a weird plot haha
image
I think you missed the ```self.sasa.sum(axis=1)``` to sum the per atom surface area. 

And also, a plot per residue per frame seems like it can be too much to see in one single plot. 
Then in the same part, when you check if there's only one frame, instead of using ```self.sasa.ndims```, better use self.traj.n_frames, which will not change after you make the reduction of the sasa matrix. 

With the same check, you can then decide to either split the plot in two or not.
image

my suggestion would be change the plot_sasa() and calculate_sasa() functions like this:

```
    def calculate_sasa(self, probe_radius=0.14):
        """
        Calculate the Solvent Accessible Surface Area (SASA) for each
        frame in the trajectory using Shrake-Rupley algorithm. Hydrogens
        are excluded for SASA residues calculation.

        Parameters:
        probe_radius (float, optional): The radius of the probe used to calculate SASA.
            Default is 0.14 nm (1.4 Å).
        """
        print("about to calculate total sasa")
        self.sasa = md.shrake_rupley(self.traj, probe_radius=probe_radius, mode="atom")
        self.sasa = self.sasa.sum(axis=1)
        print("about to calculate residue sasa")
        if self.traj.n_frames == 1:
            self.residue_sasa = md.shrake_rupley(
                self.traj, probe_radius=probe_radius, mode="residue"
            )

        # save to file
        sasa_file = f"{self.path_registry.ckpt_records}/sasa_{self.molecule_name}.csv"
        i = 0
        while os.path.exists(sasa_file):
            i += 1
            sasa_file = (
                f"{self.path_registry.ckpt_records}/sasa_{self.molecule_name}_{i}.csv"
            )
        np.savetxt(sasa_file, self.sasa, delimiter=",", header="SASA (nm²)")
        self.path_registry.map_path(
            f"sasa_{self.molecule_name}_{i}",
            sasa_file,
            description=f"SASA values for the molecule {self.molecule_name}",
        )
        return f"SASA values computed and saved to {sasa_file}"
    def plot_sasa(self):
        """
        Plot the total SASA and per-residue SASA over time.

        Returns:
        None
        """
        message = ""
        if self.sasa is None or self.residue_sasa is None:
            message += self.calculate_sasa()

        # Returns areasnp.array, shape=(n_frames, n_features)
        # if there's only one frame, don't plot
        if self.traj.n_frames == 1:
            message += " Only one frame in trajectory. No SASA plot generated."
            return message

        fig_analysis = f"sasa_{self.molecule_name}"
        fig_name = self.path_registry.write_file_name(
            type=FileType.FIGURE, fig_analysis=fig_analysis, file_format="png"
        )
        fig_id = self.path_registry.get_fileid(file_name=fig_name, type=FileType.FIGURE)
        plt.figure(figsize=(10, 5))
        if self.traj.n_frames == 1:
            plt.subplot(121)
        plt.plot(self.sasa)
        plt.xlabel("Frame")
        plt.ylabel("Total SASA (nm²)")
        plt.title("Total SASA over Time")
        if self.traj.n_frames == 1:
            plt.subplot(122)
            plt.imshow(self.residue_sasa.T, aspect="auto", interpolation="nearest")
            plt.colorbar(label="SASA (nm²)")
            plt.xlabel("Frame")
            plt.ylabel("Residue")
            plt.title("Per-residue SASA over Time")
            plt.tight_layout()
        print(f"Saving figure to {fig_name}")
        plt.savefig(f"{self.path_registry.ckpt_figures}/{fig_name}")
        plt.close()
        self.path_registry.map_path(
            fig_id,
            f"{self.path_registry.ckpt_figures}/{fig_name}",
            description=f"Plot of SASA over time for {self.molecule_name}",
        )
        message += (
            f"SASA analysis completed. Saved {fig_name}.png with plot ID {fig_id}"
        )
        return message
```

Jgmedina95:
The intertia tools seems to work alright. Just some edits on the outputs:
a) Plot of moments of inertia over time saved as: FIG_MOI_sim0_165714_20240604_174232.png.png with plot ID fig0_174232. 
You can see that the .png is repeated.
b) Also, I don't consider necessary to indicate in the output of the tool the directory of the files as the agent only interacts with IDs not with paths via path registry. 
c) On a similar note, it read better if you finish the additions to the "message" with a space, or with "\n", so it reads better. Even though this is seen mainly by the agent, it reads funny, an example: "...with file ID MOI_sim0_165714Plot of moment..."

mdagent/tools/base_tools/analysis_tools/inertia.py
```diff
@@ -0,0 +1,159 @@
...
+
+def save_to_csv(path_registry, data, file_id, description=None):
+    file_path = f"{path_registry.ckpt_records}/{file_id}.csv"
+    i = 0
+    while os.path.exists(file_path):
+        i += 1
+        file_path = f"{path_registry.ckpt_records}/{file_id}_{i}.csv"
+    np.savetxt(file_path, data, delimiter=",")
+    path_registry.map_path(file_id, file_path, description=description)
+    return file_path
```
SamCox822, line 37: We'll probably need this in several places, can you generalize them & move them to utils?

mdagent/tools/base_tools/analysis_tools/inertia.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class MomentOfInertiaToolInput(BaseModel):
+    top_fileid: str = Field(None, description="File ID for the topology file.")
+    traj_fileid: Optional[str] = Field(
+        None, description="File ID for the trajectory file."
+    )
+    molecule_name: Optional[str] = Field(
+        None, description="Name of the molecule or protein."
+    )
```
SamCox822, line 130: Not sure we need to use this for this functionality, seems overly complex.

mdagent/tools/base_tools/analysis_tools/sasa.py
```diff
@@ -0,0 +1,154 @@
...
+        return message
+
+
+class SASAToolInput(BaseModel):
+    top_fileid: str = Field(None, description="File ID for the topology file.")
+    traj_fileid: Optional[str] = Field(
+        None, description="File ID for the trajectory file."
+    )
+    mol_name: Optional[str] = Field(None, description="Name of molecule or protein.")
+
```
SamCox822, line 127: same comment as above

tests/test_analysis_tools.py
```diff
@@ -350,3 +360,170 @@ def test_process_rmsf_results(rmsd_functions, mock_plt_savefig, mock_savetxt, pl
 
     else:
         mock_plt_savefig.assert_not_called()
+
+
+################### SASA ######################
```
SamCox822, line 365: Can you separate these into separate files instead of putting these separators?

SamCox822 (changes requested):
Can you separate the files where you have the `##### header ##### ` lines

mdagent/tools/base_tools/analysis_tools/inertia.py
qcampbel: you may be actually right. I removed them

tests/test_analysis_tools.py
qcampbel: done - I organized new test files by "test_analysis_{SOME_FXNS}", so it'd be easy to find with so many test files

mdagent/tools/base_tools/analysis_tools/sasa.py
qcampbel: actually, if you have like 300 residues, it'll return 300 values. I instead computed & returned "total_sasa", which is scalar.

qcampbel:
@Jgmedina95 your suggested plot code won't work because if traj.n_frames == 1, the function will return message "Only one frame in trajectory. No SASA plot generated. " without even going through plotting lines. I hear you though. The first subplot now should show total SASA. As for 2nd subplot, I saw some publications have "Average Area vs. Residue" and use that for 2nd subplot instead of Residue over time, to be more meaningful.

Jgmedina95 (approved):
Looks Good to Me! Thanks for considering all the comments I made :)

SamCox822 (approved):

PR #136, opened 2024-05-23
updated rmsd & ppi to use mdtraj

qcampbel:


Jgmedina95:
Hi Quinny! some comments here :). 

On the RMSD tool, you calculate both the RMSD with all atoms, and the specified selection. But when the RMSD is on one frame, it only can return the all atoms RMSD. Also, it would make sense to use at least 'protein' instead of 'all', or the solvent will also be include in the calculation right?

On the RMSF. I think it would be appropriate to use 'name CA' as default instead of all to see the fluctuations per residue and not per single atom... otherwise the plot looks like this:
image

And even when I make a specific selection, the plot still looks like this:
image.

 Because if plotting both at the same time, and it doesn't really let us say which residues are moving more or less. 

On LPRMSD, I have little comments, as I don't really understand what is doing there. 

One final comment would be that the csv files are been saved in 'figures/' directory instead of 'records/'

Jgmedina95 (changes requested):
As mentioned in my comment above :)

mdagent/tools/base_tools/analysis_tools/rmsd_tools.py
```diff
@@ -2,314 +2,192 @@
...
-                "Calculating root mean square fluctuation (RMSF)...",
-                unsafe_allow_html=True,
-            )
-            return self.compute_rmsf(selection=selection, plot=plot)
-        else:
-            raise ValueError(
-                "Invalid rmsd_type. Please choose from 'rmsd', 'pairwise_rmsd', 'rmsf'"
-            )
+    if traj_fileid is None:
+        print("Warning: no trajectory file provided.")
```
SamCox822, line 20: This should either raise the warning, or give an error

mdagent/tools/base_tools/analysis_tools/rmsd_tools.py
```diff
@@ -2,314 +2,192 @@
...
+    fig_name = path_registry.write_file_name(
+        type=FileType.FIGURE,
+        fig_analysis=fig_analysis,
+        file_format="png",
+    )
+    fig_id = path_registry.get_fileid(file_name=fig_name, type=FileType.FIGURE)
+    fig_path = f"{path_registry.ckpt_figures}/{fig_name}"
+    plt.savefig(fig_path)
+    path_registry.map_path(fig_id, fig_path, description=description)
+    return fig_id, fig_path
```
SamCox822, line 51: These are functions we will use in multiple places, can you generalize them and move them to utils?

mdagent/tools/base_tools/preprocess_tools/pdb_get.py
```diff
@@ -4,6 +4,7 @@
 import streamlit as st
 from langchain.tools import BaseTool
 from rdkit import Chem
+from rdkit.Chem import AllChem
```
SamCox822, line 7:
```suggestion
from rdkit import Chem
```

mdagent/tools/base_tools/preprocess_tools/pdb_get.py
```diff
@@ -181,7 +182,7 @@ def small_molecule_pdb(self, mol_str: str) -> str:
                 m = Chem.AddHs(m)
             except Exception:
                 pass
-            Chem.AllChem.EmbedMolecule(m)
+            AllChem.EmbedMolecule(m)
```
SamCox822, line 185:
```suggestion
            Chem.AllChem.EmbedMolecule(m)
```

tests/test_analysis_tools.py
```diff
@@ -1,27 +1,21 @@
...
-import numpy as np
 import pytest
 
 from mdagent.tools.base_tools import VisFunctions
 from mdagent.tools.base_tools.analysis_tools.plot_tools import PlottingTools
 from mdagent.tools.base_tools.analysis_tools.ppi_tools import ppi_distance
-from mdagent.tools.base_tools.analysis_tools.rmsd_tools import RMSDFunctions
+from mdagent.tools.base_tools.analysis_tools.rmsd_tools import lprmsd, rmsd, rmsf
+
+#################### PLOTTING ####################
```
SamCox822, line 11: Instead, these should just be separate files

tests/test_analysis_tools.py
```diff
@@ -86,6 +80,14 @@ def test_plot_data(plotting_tools):
         assert "All plots failed due to non-numeric data." in str(excinfo.value)
 
 
+################ VISUALIZATION #################
```
SamCox822, line 83: Instead, these should just be in separate files

tests/test_analysis_tools.py
```diff
@@ -108,16 +110,9 @@ def test_create_notebook(path_to_cif, vis_fxns):
...
-@pytest.fixture
-def rmsd_functions(get_registry):
-    reg = get_registry("raw", True)
-    pdb_file = reg.get_mapped_path("ALA_123456")
-    rmsd_functions = RMSDFunctions(reg, pdb_file, "trajectory.dcd")
-    rmsd_functions.trajectory = "trajectory.dcd"
-    rmsd_functions.pdb_file = pdb_file
-    return rmsd_functions
-
+################ RMSD & PPI #################
```
SamCox822, line 113: Instead, these should just be separate files

SamCox822 (changes requested):

mdagent/tools/base_tools/analysis_tools/rmsd_tools.py
qcampbel: it's updated to a real warning when I moved to utils. Sometimes people just want to analyze PDB without trajectory file (rmsd can do this), so this won't raise an error.

mdagent/tools/base_tools/analysis_tools/rmsd_tools.py
qcampbel: done :)

tests/test_analysis_tools.py
qcampbel: moved to "test_analysis_rmsd_ppi.py" file

tests/test_analysis_tools.py
qcampbel: header removed. see my other comment about visualization & plotting tests

Jgmedina95 (approved):
Thanks for making the changes we talked about :)

mdagent/tools/base_tools/preprocess_tools/pdb_get.py
qcampbel: if I do this, it'll give an error and fail unit tests, `AttributeError: module 'rdkit.Chem' has no attribute 'AllChem'`. I thought it might be due to version differences, so I created new conda following README instructions and still got the same error. I follow the same import statement according to rdkit docs and it worked. no clue why it worked in other branches.

SamCox822: Gotcha, no worries

SamCox822 (approved):

SamCox822:
@qcampbel can you move the tests to folders? thanks!

PR #138, opened 2024-06-06
put tests into functions

SamCox822:


qcampbel:
why are we getting rid of streamlit? I thought it would be useful from user's end

qcampbel:
are unit tests for RGy missing?

SamCox822:
> are unit tests for RGy missing?

Maybe? I didn't remove any files here though so its not due to this PR

SamCox822:
> why are we getting rid of streamlit? I thought it would be useful from user's end

Maybe, but it's unusable at the moment, we'll need to redo it fully. I think we should just remove it for now until we can fix it

qcampbel (approved):
Approved. we should add RGy unit tests asap though.

PR #139, opened 2024-06-10
Bond angles dihedrals

brittyscience:
Bond angles and dihedrals code and unit test.


Jgmedina95:
Hi Brittany, thanks for working on this. 
My general comments are that this tools as they are, are hard to use, they return a big array of angles and a big array of indices... which I don't believe can be very useful, neither for the agent nor the user. It would be better to at least save the values into a file, but even then, a list of values is not very useful alone. Is not that important to wrap all the mdtraj functions as tools, but rather make general use cases from them.

Here I found some nice plots that could be done with the information from these angles(yes, Ram plots are here too!):
https://dunbrack.fccc.edu/bbdep2010/ConformationalAnalysis.php

Please try to include one or some of these plots, first. The output of the compute_chiX methods are an array of indices and an array of values per angle. 
The documentation states:
```
Returns
indicesnp.ndarray, shape=(n_chi, 4)
The indices of the atoms involved in each of the chi dihedral angles

anglesnp.ndarray, shape=(n_frames, n_chi)
The value of the dihedral angle for each of the angles in each of the frames.
```
So I imagine it may be useful to identify/map which residues correspond to the atom indices for the corresponding plots.
Something like this may be helpful
```
# Create a dictionary to map atom indices to residue names
traj = mdtraj.load(trajectory_path, top=topology_path)
atom_to_residue = {atom.index: atom.residue for atom in traj.topology.atoms}

# For each set of chi1 indices, map them to their residues
residues_per_chi1 = [[atom_to_residue[idx] for idx in chi_set] for chi_set in chi1_indices]

```

mdagent/tools/base_tools/analysis_tools/bond_angles_dihedrals_tool.py
```diff
@@ -267,4 +281,37 @@ async def _arun(self, traj_file, top_file=None):
...
+    def _run(self, traj_file, top_file=None):
+        try:
+            traj = load_single_traj(self.path_registry, traj_file, top_file)
+            if not traj:
+                return "Failed. Trajectory could not be loaded."
+
+            phi_indices, phi_angles = md.compute_phi(traj, periodic=True, opt=True)
+            psi_indices, psi_angles = md.compute_psi(traj, periodic=True, opt=True)
+
+            plt.figure(figsize=(10, 8))
```
Jgmedina95, line 304: missing to save the figure with the path registry

brittyscience:
Thank you for your feedback, Jorge. There still need to be more troubleshooting with the ram plot and the output of indices. Something is still missing, but the code is there. Hopefully it just needs a bit of tweaking.

brittyscience:
When you have time, I would really appreciate a second set of eyes to see what I am missing here. The script seems to be all set, minus the ram plot. I am missing something here. @Jgmedina95

brittyscience:
I need a second set of eyes here. More details in Notion stand up notes about this.

SamCox822 (changes requested):
see the comments I made on the other PRs, most apply to this one too. additionally, can you combine the Chi tools into 1 tool and have the agent specify the type or some other mapping? rather than chi1, chi2, etc. which will probably be difficult for the agent

will the agent ever have a preference of the types?

tests/test_preprocess/test_uniprot.py
```diff
@@ -515,6 +515,8 @@ def test_get_gene_names(query_uniprot):
     assert all(gene in all_genes for gene in specific_gene)
 
 
+# skip it because of server error
+@pytest.mark.skip(reason="Server error")
```
SamCox822, line 519: I dont think we need to skip this -> can you either let it fail or handle the error within the test?

SamCox822 (changes requested):

PR #141, opened 2024-06-14
Pdb tools

SamCox822:


mdagent/tools/base_tools/preprocess_tools/uniprot.py
```diff
@@ -0,0 +1,1250 @@
...
+                        backoff_factor=0.25,
+                        status_forcelist=[500, 502, 503, 504],
+                    )
+                ),
+            )
+            response = session.post(
+                f"{self.API_URL}/idmapping/run",
+                data={"from": from_db, "to": to_db, "ids": query},
+            )
+            response.raise_for_status()
```
Jgmedina95, line 77: Does it need handlers in case of exceptions? Or are they handle internally, and just keeps retrying?

mdagent/tools/base_tools/preprocess_tools/uniprot.py
```diff
@@ -0,0 +1,1250 @@
...
+                "https://",
+                HTTPAdapter(
+                    max_retries=Retry(
+                        total=5,
+                        backoff_factor=0.25,
+                        status_forcelist=[500, 502, 503, 504],
+                    )
+                ),
+            )
+            url = f"https://rest.uniprot.org/uniprotkb/search?fields={desired_field}&format={format_type}&query={query}"
```
Jgmedina95, line 128: whoops! bit too long 
```suggestion
            url = f"https://rest.uniprot.org/uniprotkb/search?fields={desired_field}\
            &format={format_type}&query={query}"
```

mdagent/tools/base_tools/preprocess_tools/uniprot.py
```diff
@@ -0,0 +1,1250 @@
...
+        short_names: bool = True,
+        alternative_names: bool = True,
+    ) -> list:
+        """
+        Get the protein name for a specific protein, with the option to
+        filter by primary accession number and to include alternative
+        and shortened names.
+
+        Args:
+            query: The query string to search
```
Jgmedina95, line 189:
```suggestion
            query: The query string to search (e.g. 'hemoglobin')
```

mdagent/tools/base_tools/preprocess_tools/uniprot.py
```diff
@@ -0,0 +1,1250 @@
...
+        """use the tool."""
+        try:
+            active_sites = self.uniprot.get_relevant_sites(
+                query, primary_accession, "active"
+            )
+            binding_sites = self.uniprot.get_relevant_sites(
+                query, primary_accession, "binding"
+            )
+            sites = self.uniprot.get_relevant_sites(query, primary_accession, "sites")
+            return (
```
Jgmedina95, line 925: If there are none in any of these three options can it be changed to something like: "No Binding Sites"?

mdagent/tools/maketools.py
```diff
@@ -74,6 +92,24 @@ def make_all_tools(
         SmallMolPDB(path_registry=path_instance),
```
Jgmedina95, line 92: The list of tools is starting to get big, should we include a way for the user to select which tools he wants? 
Like a list of tools strings, even for us, when testing could make it easier, as we wouldnt need to come here and comment out this part of the code

tests/test_preprocess/test_uniprot.py
```diff
@@ -0,0 +1,529 @@
...
+    assert true_binding_sites in binding_sites
+    assert not active_sites
+    assert not sites
+
+
+def test_get_protein_function(query_uniprot):
+    fxn_data_specific = query_uniprot.get_protein_function("hemoglobin", "P69905")
+    texts = [
+        "Involved in oxygen transport from the lung to the various peripheral tissues",
+        "Hemopressin acts as an antagonist peptide of the cannabinoid receptor CNR1 (PubMed:18077343). Hemopressin-binding efficiently blocks cannabinoid receptor CNR1 and subsequent signaling (PubMed:18077343)",  # noqa: E501
```
Jgmedina95, line 86: whoops! too long
```suggestion
        "Hemopressin acts as an antagonist peptide of the cannabinoid receptor CNR1 \
        (PubMed:18077343). Hemopressin-binding efficiently blocks cannabinoid receptor\
         CNR1 and subsequent signaling (PubMed:18077343)",  # noqa: E501
```

Jgmedina95 (commented):
I think is alright!
Im just a little worried by the number of total tools we're reaching.
As in one of the comments, maybe we should add an easier way to decide which tools to include from the agent initialization, maybe using their names?
something like
tool_names = get_tool_names_available()
agent = MDAgent("Structured", tools = tool_names[:10]) #to only include the first 10 tools...

Besides that, my only concern is in the first util function that doesn't have any exception handlers in case of errors.

mdagent/tools/maketools.py
SamCox822: Yeah, we can just make it an init option when initializing mdagent, specifying which tools or all

mdagent/tools/base_tools/preprocess_tools/uniprot.py
SamCox822: that's actually not true, as it returns the name lol. the input query should be another identifier

mdagent/tools/base_tools/preprocess_tools/uniprot.py
SamCox822: fixed

mdagent/tools/base_tools/preprocess_tools/uniprot.py
SamCox822: done

mdagent/tools/base_tools/preprocess_tools/uniprot.py
SamCox822: I'd rather not break it up for readability purposes since its a URL

SamCox822:
should all be fixed @Jgmedina95

Jgmedina95 (approved):

PR #144, opened 2024-06-25
changing deprecated method .simps to .simpson, and pinning scipy version

Jgmedina95:
@SamCox822 Bug fixing on scipy method


SamCox822 (approved):
lgtm thanks

PR #145, opened 2024-06-25
Removing Query Filter for now

Jgmedina95:
Based on current tests, the agent behaves way more erratically with this prompt system vs the previous design


tests/test_utils/test_query_filter.py
```diff
@@ -30,6 +41,7 @@ def test_parameters_repr():
     assert repr(params) == expected_repr
 
 
+@pytest.mark.skip(reason="We skip because query filter was moved to WIP branch.")
 def test_query_filter_simple():
     raw_query = "Find the melting point of NaCl"
     example = Example(
```
SamCox822, line 47: then lets remove

tests/test_utils/test_query_filter.py
```diff
@@ -1,7 +1,18 @@
...
+        query_filter,
+    )
+except Exception:
+    pytest.skip(allow_module_level=True)
 
 
+@pytest.mark.skip(reason="We skip because query filter was moved to WIP branch.")
 def test_parameters_repr():
     params = Parameters(
         Temperature=300,
```
SamCox822, line 18: then lets get rid of these

SamCox822 (approved):

PR #147, opened 2024-06-29
Improving Utils.get_single_traj util function

Jgmedina95:
checking if top_fileid is none, as it is usually the optional part in… tools


SamCox822 (approved):
lgtm

PR #151, opened 2024-07-09
Feature solvent forcefield inclusion

Jgmedina95:
Makes possible the use of solvents by identifying the missing solvent in the forcefield and using SmirnoffTemplateGenerator from openff

Current solvents allowed are:
        solvent_list = ["MOH", "EOH", "HOH", "ACN", "URE", "DMS", "DMF", "GOL", "BNZ"]
methanol, ethanol, water, urea,benzene, etc...


mdagent/tools/base_tools/simulation_tools/setup_and_run.py
```diff
@@ -873,40 +893,144 @@ def _create_system(
...
+                            print(f"Residue code: {residue_code}")
+                        else:
+                            print("No residue code found in the error message.")
+                            raise ValueError(str(e))
+                        if residue_code not in solvent_list:
+                            print(
+                                "Residue code not in solvent list. Adding forcefield \
+                                  not supported."
+                            )
+                            raise ValueError(str(e))
```
qcampbel, line 922: this if-else block is a bit messy and can be simplified to fewer lines, I think. For an instance, you can use "if" statement only to return all errors: `if no match: .. raise ValueError` and then `if residue_code not in solvent_list:... raise ValueError`. Then keep working on what you want to do. it'd be cleaner and easier to read logic that way.  It's called Early Return: https://medium.com/swlh/return-early-pattern-3d18a41bba8 -- and I love that funny quote in the intro fig in that article lol

qcampbel: From what you wrote here, it seems that small molecules are always treated as solvent. does that work for standalone small molecules as well? Is that what we want?

mdagent/tools/base_tools/simulation_tools/setup_and_run.py
```diff
@@ -683,7 +687,23 @@ def setup_system(self):
...
+            if "No template found for" in str(e):
+                raise ValueError(str(e))
+            else:
+                print(
+                    f"Error building system. Please check the forcefield \
+                      files {str(e)}"
+                )
+                raise ValueError(
+                    f"Error building system. Please check the forcefield files {str(e)}"
+                )
```
qcampbel, line 706: I'm not sure if I understand the purpose of this block. Basically, there's two ValueErrors, but why state "Error building system" twice, in line 700 then line 705? Is it because our MDAgent doesn't always print out the Observation step?

mdagent/tools/base_tools/simulation_tools/setup_and_run.py
```diff
@@ -873,40 +893,144 @@ def _create_system(
...
+                            smirnoff = SMIRNOFFTemplateGenerator(molecules=molecule)
+                            forcefield.registerTemplateGenerator(smirnoff.generator)
+                            attempts += 1
+                            print(
+                                f"Attempt {attempts} to add small \
+                                  molecules to forcefield."
+                            )
+                            continue
+                    else:
+                        raise ValueError(str(e))
```
qcampbel, line 1011: Looks like these lines are identical to line 905-942. If the approach to ValueError is same for both `if solvate` and `else` conditions, then what's the point of typing them out twice (~40 lines)? Can you write it once somewhere? Also, are force fields identical for standalone small molecules vs. solvent molecules? It seems like the only difference is this one line, `self.modeller.addSolvent(forcefield)`.

qcampbel (changes requested):

mdagent/tools/base_tools/simulation_tools/setup_and_run.py
Jgmedina95: is not that small molecules are always treated as solvents. Is that based on our conversation, we decided to just worked on solvents for now, and avoid protein ligands interactions. @SamCox822 right?

mdagent/tools/base_tools/simulation_tools/setup_and_run.py
Jgmedina95: oh, it can be confusing youre right. This will happen in case the residue is not part of the solvent list, then it will raise an error. The print statement before the second raise was basically debugging. Ill get rid of it :)

mdagent/tools/base_tools/simulation_tools/setup_and_run.py
Jgmedina95: the difference is that in the .addSolvent is used if the simulation is with water. The other one is if is done in vacuum. 
yeah, I could condense the workflow, but at the beginning I wasn't sure if I was going to end up using exactly the same code, or if one of the options could require a different approach

mdagent/tools/base_tools/simulation_tools/setup_and_run.py
qcampbel: protein-ligand interaction is a separate thing, for sure. I was referring to molecules only vs. molecules + proteins. As we discussed, you confirmed it can do both scenarios. I still think this block full of nested conditions (line 905-942) could be shorter and more readable lol

qcampbel (approved):
LMTG. thanks for doing this!

PR #153, opened 2024-08-06
secondary structure analysis fix

SamCox822:
- defaults to only last frame of trajectory when getting secondary structure residue counts
- better message indicating "x resides in ss" rather than just "x ss"
- fixed broken uniprot get id test
- removed stale nb experiments (rgy, rmsd, etc) that don't run anymore
- ensure that rgy figures do not overwrite each other


mdagent/tools/base_tools/analysis_tools/rgy.py
Jgmedina95, line 1: The changes do help to avoid overwriting the plots already saved. But there still an issue:
```
            self.pdb_path = [
                name
                for name in all_names
                if pdb_id in name and ".pdb" in name and "records" in name
            ][0]
``` 
and 
```
            self.dcd_path = [
                name
                for name in all_names
                if pdb_id in name and ".dcd" in name and "records" in name
            ][0]
```
Only take the first file, and the agent can only include the pdb id, so when more than one simulation is done on the pdb, the other simulations are not reachable. I would suggest either 1) performing the rgy analysis over all the matches in records, but that could be problematic on its own, or 2) instead of asking for the pdb file as an input, ask for the trajectory files too, as done with other tools, the agent seems to handle this without a problem.

mdagent/tools/base_tools/analysis_tools/secondary_structure.py
Jgmedina95, line 1: changing the description of the Tool:
“”"Compute the DSSP (secondary structure) assignment
    for a protein trajectory. Input is a trajectory file ID
    and an optional topology file ID.
    The output is an array with the DSSP code for each
    residue at each time point.“”" --->
“”"Compute the DSSP (secondary structure) assignment
    for a protein trajectory. Input is a trajectory file ID
    and an optional topology file ID.
    The output is an array with the DSSP code for each
    residue at the final frame”“”

and 
i would suggest only another change to include in the returned message that the results are based on the last frame

Jgmedina95 (commented):
See my individual comments

Jgmedina95:
Made some small changes in rgy tools, only one bug related to the path registry: It was saving the path as the id:
 ```
"/Users/jorgemedina/Desktop/md-agent/ckpt/ckpt_1/figures/radii_of_gyration_rec0_215810.csv": {
        "path": "/Users/jorgemedina/Desktop/md-agent/ckpt/ckpt_1/figures/radii_of_gyration_rec0_215810.csv",
        "name": "radii_of_gyration_rec0_215810.csv",
        "description": "Radii of gyration per frame for rec0_215810"
    }
```
Also, in case these tools are used sequentially, also included a check to see if the radii csv file was already saved, to avoid repetition on the calculation.

**Comment by Jgmedina95** [comment]
Corrected a bug related to the path registry in dssp:
```
    "rec0_110717": {
        "path": "/Users/jorgemedina/Desktop/md-agent/mdagent/dssp_20240808_110717.No file format provided",
        "name": "dssp_20240808_110717.No file format provided",
        "description": "dssp values for trajectory with id: rec0_215810"
    },
```
Now it saves as:
```
    "rec0_111801": {
        "path": "/Users/jorgemedina/Desktop/md-agent/ckpt/ckpt_1/records/dssp_20240808_111801.npy",
        "name": "dssp_20240808_111801.npy",
        "description": "dssp values for trajectory with id: rec0_215810"
    }
```
With the correct file name and path

qcampbel (approved):
may want to add a print statement if RGy file already exists and skip the calculation (Line 34 in rgy.py). Otherwise, lgtm.

PR #156, opened 2024-08-09
Clear experiments redo

SamCox822:
just clearing experiment outputs so that we can redo them -> old ones are saved elsewhere

note: the actual _prompts_ in each notebook will need to still be updated as we run it to new versions.

this is just so we aren't mixing new/old


qcampbel (approved):
lgtm. Thanks for doing this!

PR #158, opened 2024-08-19
remove chroma & simplify dependencies

qcampbel:
The purpose of this is to remove the use of Chroma VDB and reduce dependencies required to run MDAgent. Instead of using chroma, tools are vectorized either using openai embeddings or sklearn's built-in vectorization algorithm. Then, search for most similar tools is conducted using cosine similarity. 

1. Replaced tool retrieval function
2. removed unused package (outlines)
3. made some packages optional (i.e.  alternative llm providers)
4. updated README


Jgmedina95 (approved):
lgtm

PR #159, opened 2024-10-01
Update Literature Search tool

qcampbel:
* remove paperscraper. Instead, MDAgent object takes paper directory as an argument 
* upgrade to use PaperQA2


SamCox822 (approved):
lgtm

PR #161, opened 2024-11-01
various fixes

SamCox822:
- default pressure=1 
- auto-update precommit
- pqa local only -> remove unnecessary api key checks

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,7 +5,5 @@
 # OpenAI API Key
 OPENAI_API_KEY=YOUR_OPENAI_API_KEY_GOES_HERE  # pragma: allowlist secret
 
-# PQA API Key to use LiteratureSearch tool (optional) -- it also requires OpenAI key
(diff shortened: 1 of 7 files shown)


qcampbel (commented):
I read through all changes and all seem logical. While you're at it, can you also fix RGY tool, the fig paths are saved in path registry without ‘png’

SamCox822:
> I read through all changes and all seem logical. While you're at it, can you also fix
> 
> * RGY tool, the fig paths are saved in path registry without ‘png’
> * simulation tool to return a list of available forcefield files if the FF arguments are incorrect. you can find the list on top of that file, I believe
> 
> I can fix them for you later this afternoon or Monday if you're too busy

can we fix those in a separate PR since they're unrelated? I'd like to get this PR merged in ASAP as it causes a common error with claude LLMs and I'll need to rerun them with this error

qcampbel (approved):

PR #162, opened 2024-11-04
more various fixes (rgy)

SamCox822:
- rgy plots missing png 
- combined rgy tools into 1 (compute, plot & return average)
- added unit test for tool `_run`
- added more extensive error msg if ff not found


qcampbel (approved):
LGTM

PR #163, opened 2025-01-08
Claude experiments

SamCox822:


SamCox822:
@qcampbel this is ready to merge with experiments

